PR #2, opened 2022-10-20
new ELIP: Confidential Transaction descriptors

apoelstra:
This ELIP describes CT descriptors, as needed to attach blinding keys to existing descriptors and therefore create confidential addresses.

It does not cover:

* Ordinary Elements descriptors, e.g. eltr; these need a separate standard preceding this one
* Pegins or pegouts
* Anything related to PAK keys

Dupliicate of #1 created because of Github bugs.

### DIFF
--- a/elip-ct-descriptors.mediawiki
+++ b/elip-ct-descriptors.mediawiki
@@ -47,12 +47,14 @@ The current most popular scheme, [https://github.com/satoshilabs/slips/blob/mast
 We propose a new <code>ct</code> descriptor which wraps any other descriptor type in the form <code>ct(<BLINDING_KEY>, <DESCRIPTOR>)</code>.
 Here <code>DESCRIPTOR</code> refers to any existing descriptor, e.g. <code>elwsh(...)</code> or <code>eltr(...)</code> and <code>BLINDING_KEY</code> is a new expression which must be one of
 * <code>slip77(<64-character hex>)</code> which indicates that blinding keys for these addresses are derived via SLIP77; or
-* <code><KEY></code> which is an ordinary (extended) public key; the blinding key is derived by a Taproot-style tweak of the key with the data of <code><DESCRIPTOR></code>.
+* <code><KEY></code> which is a key expression as described in BIP 380, with the exception that uncompressed and x-only keys are not allowed; the blinding key is derived by a Taproot-style tweak of the key with the data of <code><DESCRIPTOR></code>.
 
 In the <code><KEY></code> expression, all key(s) must be compressed; x-only and uncompressed keys are invalid.
 
 It is permissible, and likely to be common, for a <code><KEY></code> key to be a private key (e.g. an xprv) even if the keys in the actual descriptor are public keys.
 Such a descriptor is termed a '''view descriptor''', and the blinding private key a '''view key'''.
+
+Later extensions to this ELIP may specify alternate expressions for `KEY`, for example to allow MuSig-combining keys from multiple participants.
 
 ===Drawbacks===
 
@@ -95,19 +97,19 @@ Using the <code>ct(slip77(<64-byte hex>), <DESCRIPTOR>)</code> construction, any
 The following CT descriptors should be parseable and generate the given addresses:
 
 * Valid Descriptor 1: <code>ct(xpub6ERApfZwUNrhLCkDtcHTcxd75RbzS1ed54G1LkBUHQVHQKqhMkhgbmJbZRkrgZw4koxb5JaHWkY4ALHY2grBGRjaDMzQLcgJvLJuZZvRcEL,elpkh(xpub69H7F5d8KSRgmmdJg2KhpAK8SR3DjMwAdkxj3ZuxV27CprR9LgpeyGmXUbC6wb7ERfvrnKZjXoUmmDznezpbZb7ap6r1D3tgFxHmwMkQTPH))#y0lg3d5y</code>
-** Blinding public key: <code>xpub6ERApfZwUNrhLCkDtcHTcxd75RbzS1ed54G1LkBUHQVHQKqhMkhgbmJbZRkrgZw4koxb5JaHWkY4ALHY2grBGRjaDMzQLcgJvLJuZZvRcEL</code>
+** Descriptor blinding public key: <code>xpub6ERApfZwUNrhLCkDtcHTcxd75RbzS1ed54G1LkBUHQVHQKqhMkhgbmJbZRkrgZw4koxb5JaHWkY4ALHY2grBGRjaDMzQLcgJvLJuZZvRcEL</code>
 ** Confidential address: <code>CTEnDa5fqGccV3g3jvp4exSQwRfb6FpGchNBF4ZrAaq8ip8gvLqHCtzw1F7d7U5gYJYXBwymgEMmJjca</code>
 ** Unconfidential address: <code>2dhfebpgPWpeqPdCMMam5F2UHAgx3bbLzAg</code>
 * Valid Descriptor 2: <code>ct(xpub6ERApfZwUNrhLCkDtcHTcxd75RbzS1ed54G1LkBUHQVHQKqhMkhgbmJbZRkrgZw4koxb5JaHWkY4ALHY2grBGRjaDMzQLcgJvLJuZZvRcEL,elwpkh(xpub69H7F5d8KSRgmmdJg2KhpAK8SR3DjMwAdkxj3ZuxV27CprR9LgpeyGmXUbC6wb7ERfvrnKZjXoUmmDznezpbZb7ap6r1D3tgFxHmwMkQTPH))#kt4e25qt</code>
-** Blinding public key: <code>xpub6ERApfZwUNrhLCkDtcHTcxd75RbzS1ed54G1LkBUHQVHQKqhMkhgbmJbZRkrgZw4koxb5JaHWkY4ALHY2grBGRjaDMzQLcgJvLJuZZvRcEL</code>
+** Descriptor blinding public key: <code>xpub6ERApfZwUNrhLCkDtcHTcxd75RbzS1ed54G1LkBUHQVHQKqhMkhgbmJbZRkrgZw4koxb5JaHWkY4ALHY2grBGRjaDMzQLcgJvLJuZZvRcEL</code>
 ** Confidential address: <code>el1qqg5s7xj7upzl7h4q2k2wj4vq63nvaktn0egqu09nqcr6d44p4evaqknpl78t02k2xqgdh9ltmfmpy9ssk7qfvrldr2dttt3ez</code>
 ** Unconfidential address: <code>ert1qtfsllr4h4t9rqyxmjl4a5asjzcgt0qyk32h3ur</code>
 * Valid Descriptor 3: <code>ct(xpub6ERApfZwUNrhLCkDtcHTcxd75RbzS1ed54G1LkBUHQVHQKqhMkhgbmJbZRkrgZw4koxb5JaHWkY4ALHY2grBGRjaDMzQLcgJvLJuZZvRcEL,elsh(wpkh(xpub69H7F5d8KSRgmmdJg2KhpAK8SR3DjMwAdkxj3ZuxV27CprR9LgpeyGmXUbC6wb7ERfvrnKZjXoUmmDznezpbZb7ap6r1D3tgFxHmwMkQTPH)))#xg9r4jej</code>
-** Blinding public key: <code>xpub6ERApfZwUNrhLCkDtcHTcxd75RbzS1ed54G1LkBUHQVHQKqhMkhgbmJbZRkrgZw4koxb5JaHWkY4ALHY2grBGRjaDMzQLcgJvLJuZZvRcEL</code>
+** Descriptor blinding public key: <code>xpub6ERApfZwUNrhLCkDtcHTcxd75RbzS1ed54G1LkBUHQVHQKqhMkhgbmJbZRkrgZw4koxb5JaHWkY4ALHY2grBGRjaDMzQLcgJvLJuZZvRcEL</code>
 ** Confidential address: <code>AzpnREsN1RSi4JB7rAfpywmPsvGxyygmwm9o3iZcP43svg4frVW5DXvGj5yEx6mKcPtAyHgQWVikFRCM</code>
 ** Unconfidential address: <code>XKGUGskfGsNRR1Ww4ytemgBjuszohUaNgv</code>
 * Valid Descriptor 4: <code>ct(xpub6ERApfZwUNrhLCkDtcHTcxd75RbzS1ed54G1LkBUHQVHQKqhMkhgbmJbZRkrgZw4koxb5JaHWkY4ALHY2grBGRjaDMzQLcgJvLJuZZvRcEL,eltr(xpub69H7F5d8KSRgmmdJg2KhpAK8SR3DjMwAdkxj3ZuxV27CprR9LgpeyGmXUbC6wb7ERfvrnKZjXoUmmDznezpbZb7ap6r1D3tgFxHmwMkQTPH))#c0pjjxyw</code>
-** Blinding public key: <code>xpub6ERApfZwUNrhLCkDtcHTcxd75RbzS1ed54G1LkBUHQVHQKqhMkhgbmJbZRkrgZw4koxb5JaHWkY4ALHY2grBGRjaDMzQLcgJvLJuZZvRcEL</code>
+** Descriptor blinding public key: <code>xpub6ERApfZwUNrhLCkDtcHTcxd75RbzS1ed54G1LkBUHQVHQKqhMkhgbmJbZRkrgZw4koxb5JaHWkY4ALHY2grBGRjaDMzQLcgJvLJuZZvRcEL</code>
 ** Confidential address: <code>el1pq0nsl8du3gsuk7r90sgm78259mmv6mt9d4yvj30zr3u052ufs5meuc2tuvwx7k7g9kvhhpux07vqpm3qjj8uwdj94650265ustv0xy8z2pc847zht4k0</code>
 ** Unconfidential address: <code>ert1pv997x8r0t0yzmxtms7r8lxqqacsffr78xez6a284d2wg9k8nzr3q3s6527</code>
 * Valid Descriptor 5: <code>ct(slip77(b2396b3ee20509cdb64fe24180a14a72dbd671728eaa49bac69d2bdecb5f5a04),elpkh(xpub69H7F5d8KSRgmmdJg2KhpAK8SR3DjMwAdkxj3ZuxV27CprR9LgpeyGmXUbC6wb7ERfvrnKZjXoUmmDznezpbZb7ap6r1D3tgFxHmwMkQTPH))#hw2glz99</code>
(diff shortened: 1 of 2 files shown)


elip-ct-descriptors.mediawiki
```diff
@@ -0,0 +1,98 @@
...
+The current most popular scheme, [https://github.com/satoshilabs/slips/blob/master/slip-0077.md SLIP-77], does not satisfy any of these criteria, which limits its usefulness as we move toward a descriptor-centric setting in which multiparty addresses are common.
+
+==Design==
+
+===Overview===
+We propose a new `ct` descriptor which wraps any other descriptor type in the form `ct(<BLINDING_KEY>, <DESCRIPTOR>)`.
+Here `DESCRIPTOR` refers to any existing descriptor, e.g. `wsh(...)` or `tr(...)` and `BLINDING_KEY` is a new expression which must be one of
+* `slip77(<64-character hex>)` which indicates that blinding keys for these addresses are derived via SLIP77; or
+* `hash_to_private(<KEY_1>, <KEY_2>, ...)` which indicates that the blinding key is produced by a hash of the sorted list of one or more public keys; or
+* `<KEY>` which is an ordinary (extended) public key which will be used as a blinding key
```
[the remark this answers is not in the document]
apoelstra: Yes, it means an xpub.

apoelstra:
@LeoComandini thanks for your review! I have fixed the formatting (I think/hope), and changed the "non-concrete descriptor" text to be clearer.

elip-ct-descriptors.mediawiki
```diff
@@ -0,0 +1,96 @@
...
+* In multiparty settings, each wallet should be able to restrict this granularity
+* It should be possible somehow to do public derivation of CT addresses, given only a descriptor containing (extended) public keys
+
+The current most popular scheme, [https://github.com/satoshilabs/slips/blob/master/slip-0077.md SLIP-77], does not satisfy any of these criteria, which limits its usefulness as we move toward a descriptor-centric setting in which multiparty addresses are common.
+
+==Design==
+
+===Overview===
+We propose a new <code>ct</code> descriptor which wraps any other descriptor type in the form <code>ct(<BLINDING_KEY>, <DESCRIPTOR>)</code>.
+Here <code>DESCRIPTOR</code> refers to any existing descriptor, e.g. <code>wsh(...)</code> or <code>tr(...)</code> and <code>BLINDING_KEY</code> is a new expression which must be one of
```
LeoComandini, line 48: In the reference implementation in elemenets-miniscipt those are "elwsh", "eltr", ..., either here or there they should be changed.

elip-ct-descriptors.mediawiki
```diff
@@ -0,0 +1,98 @@
...
+
+==Design==
+
+===Overview===
+We propose a new <code>ct</code> descriptor which wraps any other descriptor type in the form <code>ct(<BLINDING_KEY>, <DESCRIPTOR>)</code>.
+Here <code>DESCRIPTOR</code> refers to any existing descriptor, e.g. <code>elwsh(...)</code> or <code>eltr(...)</code> and <code>BLINDING_KEY</code> is a new expression which must be one of
+* <code>slip77(<64-character hex>)</code> which indicates that blinding keys for these addresses are derived via SLIP77; or
+* <code><KEY></code> which is an ordinary (extended) public key; the blinding key is derived by a Taproot-style tweak of the key with the data of <code><DESCRIPTOR></code>.
+
+In both the <code><KEY></code> expression, all key(s) must be compressed; x-only and uncompressed keys are invalid.
```
LeoComandini, line 52: nit: 

> both

now we have single variant with `<KEY>`

elip-ct-descriptors.mediawiki
```diff
@@ -0,0 +1,98 @@
...
+
+The current most popular scheme, [https://github.com/satoshilabs/slips/blob/master/slip-0077.md SLIP-77], does not satisfy any of these criteria, which limits its usefulness as we move toward a descriptor-centric setting in which multiparty addresses are common.
+
+==Design==
+
+===Overview===
+We propose a new <code>ct</code> descriptor which wraps any other descriptor type in the form <code>ct(<BLINDING_KEY>, <DESCRIPTOR>)</code>.
+Here <code>DESCRIPTOR</code> refers to any existing descriptor, e.g. <code>elwsh(...)</code> or <code>eltr(...)</code> and <code>BLINDING_KEY</code> is a new expression which must be one of
+* <code>slip77(<64-character hex>)</code> which indicates that blinding keys for these addresses are derived via SLIP77; or
+* <code><KEY></code> which is an ordinary (extended) public key; the blinding key is derived by a Taproot-style tweak of the key with the data of <code><DESCRIPTOR></code>.
```
LeoComandini, line 50: > extended

Why a wallet should use a xpub/xprv as a bare blinding key?

Now that the key is tweaked with the scriptpubkey (which in most cases should be derived from some xpubs), 
I can't think of a reason why a wallet should use xpub/xprv as "master" blinding key rather than a public/private key.

Does it make sense to disallow extended keys for the "bare" variant?

apoelstra: I don't think so. We definitely should support xpubs/xprivs so that they can specify things that have BIP32 paths starting from an existing master seed (otherwise they will need to do a bunch of extra work to derive a public key and then remember how they got it).

I agree that supporting **ranged** xpubs no longer has a clear utility ... but I see no reason to ban it either.

apoelstra:
Rebased, fixed `both` nit, added test vectors.

elip-ct-descriptors.mediawiki
```diff
@@ -0,0 +1,151 @@
...
+The <code>scriptPubKey</code> corresponding to a <code>ct</code> descriptor is that corresponding to the embedded <code>DESCRIPTOR</code>. The encoding and semantics of <code>BLINDING_KEY</code> are given below.
+
+<code>BLINDING_KEY</code> expressions are one of
+* '''slip77''', encoded as <code>slip77(<64-characters hex>)</code>, whose semantics are that the 64 hex characters are interpreted as the 32-byte <code>master_blinding_key</code> in SLIP77; the <code>scriptPubKey</code> for SLIP77 is computed as normal from the ordinary descriptor.
+** These 32 bytes are the same for both public and private descriptors; they have no "public" equivalent.
+** This mode is not recommended because there is no way to express this descriptor without revealing the SLIP77 key, which can be used to unblind every single output received by the wallet.
+* '''bare key''', which simply encodes a single key in the same way as the keys in <code><DESCRIPTOR></code> are encoded. The blinding key is derived from the address by taking the public key ''K'' and tweaking it to form a new key ''K<sub>blind</sub>'' as follows:
+** ''K' = K + H<sub>tag</sub>(K, <code>scriptPubKey</code>)'', where
```
LeoComandini, line 75: Should this be `H<sub>tag</sub>(K, <code>scriptPubKey</code>) * G`? or `K` is a private key?

elip-ct-descriptors.mediawiki
```diff
@@ -0,0 +1,151 @@
...
+** SLIP77 master blinding key: <code>b2396b3ee20509cdb64fe24180a14a72dbd671728eaa49bac69d2bdecb5f5a04</code>
+** Confidential address: <code>AzprjJt3poXAJWLmanTHYB2zrkUMXiiVyCRhDS7VRRaXUsRqeCMi3vKe4YufizpYDyzvQFLsvjfMeLMD</code>
+** Unconfidential address: <code>XQ7ffnJkhMwj1H8Ma6N1vcU9mqAa96wB9w</code>
+* Valid Descriptor 8: <code>ct(slip77(b2396b3ee20509cdb64fe24180a14a72dbd671728eaa49bac69d2bdecb5f5a04),eltr(03774eec7a3d550d18e9f89414152025b3b0ad6a342b19481f702d843cff06dfc4))#65xky8c4</code>
+** SLIP77 master blinding key: <code>b2396b3ee20509cdb64fe24180a14a72dbd671728eaa49bac69d2bdecb5f5a04</code>
+** Confidential address: <code>el1pqgkj53t6cpqytk65s7dygws4a6hny3rrev7rw3r5gl7ymnjdqtt9k6r5jd0rkpzukq6hd965kepcmwtxvg0fh4ak4f636gv25yky23cehlfsszhd84mc</code>
+** Unconfidential address: <code>ert1pdp6fxh3mq3wtqdtkja2tvsudh9nxy85m67m25agayx92ztz9guvs9wr5lg</code>
+* Invalid Descriptor 1
+** <code>ct(slip77(b2396b3ee20509cdb64fe24180a14a72dbd671728eaa49bac69d2bdecb5f5a04),elsh(wpkh(03774eec7a3d550d18e9f89414152025b3b0ad6a342b19481f702d843cff06dfc4)))#xxxxxxxx</code>
+** Reason:
```
LeoComandini, line 131: nit: missing reason here and below

elip-ct-descriptors.mediawiki
```diff
@@ -0,0 +1,151 @@
...
+* Valid Descriptor 2: <code>ct(02dce16018bbbb8e36de7b394df5b5166e9adb7498be7d881a85a09aeecf76b623,elwpkh(03774eec7a3d550d18e9f89414152025b3b0ad6a342b19481f702d843cff06dfc4))#h5e0p6m9</code>
+** Blinding key: <code>02dce16018bbbb8e36de7b394df5b5166e9adb7498be7d881a85a09aeecf76b623</code>
+** Confidential address: <code>el1qq0r6pegudzm0tzpszelc34qjln4fdxawgwmgnza63wwpzdy6jrm0grmqvvk2ce5ksnxcs9ecgtnryt7xg34060uctupg60d02</code>
+** Unconfidential address: <code>ert1qpasxxt9vv6tgfnvgzuuy9e3j9lryg6ha53x9q0</code>
+* Valid Descriptor 3: <code>ct(02dce16018bbbb8e36de7b394df5b5166e9adb7498be7d881a85a09aeecf76b623,elsh(wpkh(03774eec7a3d550d18e9f89414152025b3b0ad6a342b19481f702d843cff06dfc4)))#3kvhe0a8</code>
+** Blinding key: <code>02dce16018bbbb8e36de7b394df5b5166e9adb7498be7d881a85a09aeecf76b623</code>
+** Confidential address: <code>AzpsK7uqP1KVEMfDQvBXYUkpHmFagD3W4vaLe1X7uy8MS6nj41kNYnaexuXgx14PcbNnYAqBdCSWcbga</code>
+** Unconfidential address: <code>XQ7ffnJkhMwj1H8Ma6N1vcU9mqAa96wB9w</code>
+* Valid Descriptor 4: <code>ct(02dce16018bbbb8e36de7b394df5b5166e9adb7498be7d881a85a09aeecf76b623,eltr(03774eec7a3d550d18e9f89414152025b3b0ad6a342b19481f702d843cff06dfc4))#ytq9w7f3</code>
```
LeoComandini, line 109: Can we please have a valid test vector for the `bare` variant specifying a private key?
That should be the most practical use case, i.e. allowing unblinding

elip-ct-descriptors.mediawiki
apoelstra: oops, I was supposed to fill those in by hand :P

elip-ct-descriptors.mediawiki
apoelstra: Sure, though I'd need to implement that :P. Give me a day or two. I think I can do it before https://github.com/ElementsProject/rust-elements/pull/166 (which is blocking the reference implementation) gets in.

elip-ct-descriptors.mediawiki
jgriffiths: > <code><KEY></code> which is an ordinary (extended) public key;

This should probably read something like:

`<code><KEY></code> which is an ordinary (or extended) private or public key;`

An example bare key with xpub/xprv would be useful in the examples, as would a raw private rather than public key.

apoelstra:
Sorry for the long delay. Switching to xpubs was easy but the view key support was pretty annoying because our elements-miniscript implementation doesn't directly relate secret key types to public key types, making it difficult to elegantly have "one key private, others public" for view keys. In the end I hacked it in, but as implemented it does not support e.g. ranged blinding keys.

I have updated the test vectors in the ELIP, and the reference implementation at https://github.com/ElementsProject/elements-miniscript/pull/31

elip-ct-descriptors.mediawiki
```diff
@@ -0,0 +1,167 @@
...
+This reduces implementation complexity and the need to store additional keys, at the cost of turning the descriptor unconditionally into a view descriptor.
+In other words, users of this variant will be unable to provide external parties the ability to derive addresses without ''also'' providing them the ability to unblind coins.
+
+In a future ELIP we will specify a standard deterministic key generation scheme for this case.
+
+==Backwards Compatibility==
+
+Using the <code>ct(slip77(<64-byte hex>), <DESCRIPTOR>)</code> construction, any wallet should be able to express its existing confidential addresses using this new scheme.
+
+==Test Vectors==
```
LeoComandini, line 93: Can we add test vectors with non-extended public and private blinding keys?

elip-ct-descriptors.mediawiki
```diff
@@ -0,0 +1,167 @@
...
+
+The following CT descriptors should be parseable and generate the given addresses:
+
+* Valid Descriptor 1: <code>ct(xpub6ERApfZwUNrhLCkDtcHTcxd75RbzS1ed54G1LkBUHQVHQKqhMkhgbmJbZRkrgZw4koxb5JaHWkY4ALHY2grBGRjaDMzQLcgJvLJuZZvRcEL,elpkh(xpub69H7F5d8KSRgmmdJg2KhpAK8SR3DjMwAdkxj3ZuxV27CprR9LgpeyGmXUbC6wb7ERfvrnKZjXoUmmDznezpbZb7ap6r1D3tgFxHmwMkQTPH))#y0lg3d5y</code>
+** Blinding public key: <code>xpub6ERApfZwUNrhLCkDtcHTcxd75RbzS1ed54G1LkBUHQVHQKqhMkhgbmJbZRkrgZw4koxb5JaHWkY4ALHY2grBGRjaDMzQLcgJvLJuZZvRcEL</code>
```
LeoComandini, line 98: nit: the name here is slightly confusing, "blinding public key" is the one encoded in the confidential address, this is more like the "SLIP77 master blinding key", perhaps we should name it "master blinding public key" or "descriptor blinding public key". But any name is fine for me, as long we have different one wrt the derived (public) blinding keys

apoelstra: Oh, good point. I think I'll change this to "Descriptor blinding [public] key". Agreed it's confusing now that the key in the descriptor is different from the key in the address.

elip-ct-descriptors.mediawiki
```diff
@@ -0,0 +1,167 @@
...
+
+==Design==
+
+===Overview===
+We propose a new <code>ct</code> descriptor which wraps any other descriptor type in the form <code>ct(<BLINDING_KEY>, <DESCRIPTOR>)</code>.
+Here <code>DESCRIPTOR</code> refers to any existing descriptor, e.g. <code>elwsh(...)</code> or <code>eltr(...)</code> and <code>BLINDING_KEY</code> is a new expression which must be one of
+* <code>slip77(<64-character hex>)</code> which indicates that blinding keys for these addresses are derived via SLIP77; or
+* <code><KEY></code> which is an ordinary (extended) public key; the blinding key is derived by a Taproot-style tweak of the key with the data of <code><DESCRIPTOR></code>.
+
+In the <code><KEY></code> expression, all key(s) must be compressed; x-only and uncompressed keys are invalid.
```
sanket1729, line 52: nit: It would be great if this specifically mentions `KEY` expression in descriptor spec. Reading the words compressed ordinary key confused me a bit.
 
another nit: This also mentions multi-path KEY expressions. I don't think we want to support that.

sanket1729 (approved):
ACK 530cbf46089859524eb7c77f55f13e6cc0d1c6fc. Sorry for my confusion on the PR review.

elip-ct-descriptors.mediawiki
apoelstra: Where does it mention multi-path KEY expressions? I did not even know those existed when I was writing this.

apoelstra:
I apologize for not doing the test vectors yet. I will try to get them before the end of this week.

Meanwhile we should really give this a number. I propose 150. If nobody complains I will assign this and update the PR tomorrow.

elip-ct-descriptors.mediawiki
real-or-random: Perhaps the "all key(s)" is confusing. There's just a single key.

apoelstra: It might be a `musig` of multiple keys.

real-or-random: in the `<KEY>` expression?

sanket1729: Key expressions are defined as follows. `https://github.com/bitcoin/bips/blob/master/bip-0380.mediawiki#key-expressions`. I think what we mean to say that all `KEY` expressions that are not uncompressed keys are allowed.

real-or-random: I'm not sure. We say `<code><KEY></code> which is an ordinary (extended) public key`. We don't state anywhere that this is a "key expression" as in BIP380. 

If it's supposed to be a key expression as in BIP380, we should say this. If not, we should consider renaming `KEY`.

apoelstra: Ok, I will reword this. Yes, it is supposed to be a key expression.

apoelstra: Though it would be nice if we had our own ELIP for key expressions so that we can extend it to support `musig()` eventually. Though I have zero interest in doing that. This ELIP has been more than enough standards work for one lifetime.

elip-ct-descriptors.mediawiki
```diff
@@ -0,0 +1,167 @@
...
+* In multiparty settings, each wallet should be able to restrict this granularity
+* It should be possible somehow to do public derivation of CT addresses, given only a descriptor containing (extended) public keys
+
+The current most popular scheme, [https://github.com/satoshilabs/slips/blob/master/slip-0077.md SLIP-77], does not satisfy any of these criteria, which limits its usefulness as we move toward a descriptor-centric setting in which multiparty addresses are common.
+
+==Design==
+
+===Overview===
+We propose a new <code>ct</code> descriptor which wraps any other descriptor type in the form <code>ct(<BLINDING_KEY>, <DESCRIPTOR>)</code>.
+Here <code>DESCRIPTOR</code> refers to any existing descriptor, e.g. <code>elwsh(...)</code> or <code>eltr(...)</code> and <code>BLINDING_KEY</code> is a new expression which must be one of
```
delta1, line 48:
```suggestion
Here <code>DESCRIPTOR</code> refers to any existing descriptor, e.g. <code>elwsh(...)</code> or <code>eltr(...)</code> 
```

Is this recursive - can a `ct` descriptor include another `ct` descriptor?

delta1: Cool, might be worth specifying

apoelstra: How can I be more specific than "any existing descriptor"?

sanket1729 (approved):
ACK 5726891c0b328ac0a9fbd81bd5737a0d94489eb9 . One small missing thing is that test vectors don't mention the address was created for elements regtest network. 

```rust
    pub const ELEMENTS: AddressParams = AddressParams {
        p2pkh_prefix: 235,
        p2sh_prefix: 75,
        blinded_prefix: 4,
        bech_hrp: "ert",
        blech_hrp: "el",
    };
```

apoelstra:
Thanks! Going to merge this. The test network thing is somewhat implicit by the prefix of the addresses, but we should mention it nonetheless. Can do this in a followup PR.

PR #4, opened 2022-12-01
Hardware wallet extensions for PSET

miketlk:
This is a proposal for the new ELIP specifying additional PSET fields useful for integration with hardware wallets or similar airgapped devices. By implementing the Signer role, these devices by design have no access to the network, and thus solely rely on the information present in PSET. Particularly, including asset metadata like ticker and precision is crucial for displaying adequate information to the users on the transaction they are about to sign.

### DIFF
--- a/elip-0100.mediawiki
+++ b/elip-0100.mediawiki
@@ -1,5 +1,5 @@
 <pre>
-  ELIP: ????
+  ELIP: 100
   Layer: Applications
   Title: Hardware Wallet Extensions for Partially Signed Elements Transaction Format
   Author: Mikhail Tolkachev <contact@miketolkachev.dev>
--- a/elip-pset-hww-extensions.mediawiki
+++ b/elip-pset-hww-extensions.mediawiki
@@ -62,7 +62,7 @@ For all Elements PSBT proprietary fields, PSETs cannot be combined when there ar
 ==Roles==
 
 Using the transaction format involves many different responsibilities. Multiple responsibilities can be handled by a single entity, but each responsibility is specialized in what it should be capable of doing.
-This specification relies on definition of roles in the main PSET standard. The following statemends describe specifics of handling Hardware Wallet Extensions for some of the roles for which it's applicable.
+This specification relies on definition of roles in the main PSET standard. The following statemends describe specifics of handling Hardware Wallet Extensions for some of the roles for which it is applicable.
 
 ===Updater===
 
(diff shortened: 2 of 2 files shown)


philippem:
The requirement to communicate asset metadata in the PSET is probably not unique to hardware wallets. Maybe the field name could be generalized and the applicability of the ELIP expanded?

@apoelstra thoughts?

apoelstra:
I think we need to step back and take a holistic view of PSET, what workflows it should support, etc., and come up with a coherent plan for all the fields that it should support. I'm happy to review this and give it an ELIP number, but I don't have the view of how we're planning to use this, or the time, to say whether this proposal is something we want to implement in Elements.

We've added several fields over the last fiew months based on direct requests from users to add the fields, but I don't think this sort of ad-hoc extension strategy is sustainable. (I'm not saying that's what this is! But I'm saying that I can't make a judgement call.)

apoelstra:
It looks like this is missing a Backward Compatibility section.

miketlk:
I don't think it is necessary for Elements to implement these extensions. It would be more logical to support them at the level of client software, such as Blockstream Green. Although support in Elements would be convenient for automated generation of test data for HWW app, as I do now using RPC calls.

miketlk:
> It looks like this is missing a Backward Compatibility section.

@apoelstra, thank you for looking through the draft. I will add Backward Compatibility section. It seems to me it should be trivial because the only new field is optional. Existing implementations should ignore it, as they do for each field with unknown key. The newer implementations, for which these extensions matter, should not rely on the presence of this field.

apoelstra:
> I will add Backward Compatibility section. It seems to me it should be trivial because the only new field is optional.

Yep, that's my impression as well :). But I'd still like a sentence or two saying this, then I'm happy to assign this an ELIP number (unless @tiero jumps in to say that I shouldn't be assigning numbers since I'm not listed in ELIP2 as a maintainer :))

miketlk:
@apoelstra, please check commit [bb1f245](https://github.com/ElementsProject/ELIPs/pull/4/commits/bb1f245a7070385dfb953668a5b76ec107b1abe0)

tiero:
@apoelstra  I would love to have you co-editor as well (I proposed back in April to try to push to have this venue, more people can be involved, the better)

apoelstra:
@miketlk looks good to me.

Let's give this ELIP 100 (we'll reserve the <100 numbers for "process" things). Could you add a commit which moves the file to the new name and updates the header?

miketlk:
Ok. Done in [06407a1](https://github.com/ElementsProject/ELIPs/pull/4/commits/06407a13ac10e28b68fd2e5875397e0a2cc8245e)

apoelstra (approved):
ACK 06407a13ac10e28b68fd2e5875397e0a2cc8245e

PR #8, opened 2023-11-08
Deterministic Descriptor Blinding Keys ELIP 

LeoComandini:


elip-deterministic-blinding-key.mediawiki
```diff
@@ -0,0 +1,84 @@
...
+</pre>
+
+Is equivalent to
+
+<pre>
+ct(<KEY>,<DESCRIPTOR>)
+</pre>
+
+Where <code>KEY</code> is a view blinding key derived as follow:
+* If <code>DESCRIPTOR</code> is multipath, get the first descriptor
```
RCasatta, line 72: I would consider concatenating every scriptPubKey coming from the multipath expansion to avoid collisions.

elip-deterministic-blinding-key.mediawiki
```diff
@@ -0,0 +1,84 @@
...
+* If it has wildcards, get the definite desciptor at index 0
+* Derive the <code>scriptPubKey</code> from the definite descriptor
```
RCasatta, line 74: Maybe adding a small explanation on why using the derived scriptPubKey instead of the descriptor itself, because multiple equivalent descriptor may have different representation (for example non having the fingerprint) but producing the same spk

LeoComandini: Added footnote

elip-deterministic-blinding-key.mediawiki
LeoComandini: IMHO it's not worth it, added a footnote with a detailed rationale

apoelstra: I think we ought to concatenate them, since there will always be a small finite number. (I know it's extra code but it's not *that* bad, and closes a theoretical complaint.)

In any case, I'm not really convinced by the stated rationale, which is basically "I'm not aware of anybody using this in a way that'll cause problems". But I do think it's true that, in practice, users are never going to have multiple descriptors that are identical except for having different multi-path branches.

LeoComandini: Changed to use the concatenated scriptpubkeys

elip-deterministic-blinding-key.mediawiki
```diff
@@ -0,0 +1,122 @@
...
+Here we use the <code>scriptPubKey</code> because descriptors have fields that might change without affecting the spending conditions.
+For instance changing the fingerprint of one xpub yields the same <code>scriptPubKey</code>s.
+Using the <code>scriptPubKey</code> allows us to treat such descriptors as equivalent.
+</ref>
+* Concatenate all <code>scriptPubKey</code>s obtained before.
+* Let ''H<sub>tag</sub>'' be a [https://github.com/bitcoin/bips/blob/master/bip-0340.mediawiki#Design BIP-340 tagged hash] with tag <code>Deterministic-View-Key/1.0</code><ref>
+A tagged hash is used to make sure hashes used in this context can't be reinterpreted in another one.
+</ref>
+* The descriptor blinding key is the tagged hash of the concatenated <code>scriptPubKey</code>s, ''<code>KEY</code> = H<sub>tag</sub>(<code>scriptPubKey</code>, ...) mod n'', converted to WIF format.
+
```
jgriffiths, line 95: I think you should specify here that the resulting key is a `bare key` as described in ELIP 0150, and must be tweaked according to those rules before being used for a given address. In particular, the key is not to be used as a SLIP77 style key.

The test vectors should give a solution for an arbitrary wildcard index with the resulting tweaked blinding key and address generated so implementations can verify their behaviour in this regard.

elip-deterministic-blinding-key.mediawiki
```diff
@@ -0,0 +1,122 @@
...
+==Test Vectors==
+
+The following ordinary descriptor yield to the given descriptor blinding key:
+
+* Test vector: 1
+** Ordinary descriptor: elwpkh(xpub661MyMwAqRbcFkPHucMnrGNzDwb6teAX1RbKQmqtEF8kK3Z7LZ59qafCjB9eCRLiTVG3uxBxgKvRgbubRhqSKXnGGb1aoaqLrpMBDrVxga8/<0;1>/*)#lg37z46y
+** Confidential descriptor: ct(L5QWfEraRqKYPA71SSW6cZtsSie5kCLxg2NgjgfRjzp4sewoDtLt,elwpkh(xpub661MyMwAqRbcFkPHucMnrGNzDwb6teAX1RbKQmqtEF8kK3Z7LZ59qafCjB9eCRLiTVG3uxBxgKvRgbubRhqSKXnGGb1aoaqLrpMBDrVxga8/<0;1>/*))#psv7fun2
```
jgriffiths, line 102: Once an ELIP number is assigned, the ordinary descriptors should be replaced with the new `elipXXXX` expressions.

elip-deterministic-blinding-key.mediawiki
```diff
@@ -0,0 +1,122 @@
...
+* If the top-level script expression is <code>combo</code>, fail.
+* Future extensions that support multiple descriptors other than multi-path are not supported.
+* If <code>DESCRIPTOR</code> is multi-path, expand to every single descriptor.
+* For each single descriptor:
+** If it has wildcards, get the definite descriptor at index 2<sup>31</sup>-1<ref>
+We want to use a <code>scriptPubKey</code> that will not be used in a transaction.
+Otherwise it can be used directly by anyone to derive the descriptor blinding key.
+Wallets use <code>scriptPubKey</code>s starting from index 0,
+thus we choose the last non-hardened index,
+which in practice will never be used in a transaction.
```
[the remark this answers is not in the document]
LeoComandini: It is already, since it's wrapped in between `<ref></ref>`, see https://github.com/LeoComandini/ELIPs/blob/2023-11-deterministic-blinding-key/elip-deterministic-blinding-key.mediawiki

elip-deterministic-blinding-key.mediawiki
```diff
@@ -0,0 +1,122 @@
...
+</ref>
+** Derive the <code>scriptPubKey</code> from the definite descriptor<ref>
+Here we use the <code>scriptPubKey</code> because descriptors have fields that might change without affecting the spending conditions.
+For instance changing the fingerprint of one xpub yields the same <code>scriptPubKey</code>s.
+Using the <code>scriptPubKey</code> allows us to treat such descriptors as equivalent.
+</ref>
+* Concatenate all <code>scriptPubKey</code>s obtained before.
+* Let ''H<sub>tag</sub>'' be a [https://github.com/bitcoin/bips/blob/master/bip-0340.mediawiki#Design BIP-340 tagged hash] with tag <code>Deterministic-View-Key/1.0</code><ref>
+A tagged hash is used to make sure hashes used in this context can't be reinterpreted in another one.
```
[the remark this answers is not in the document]
LeoComandini: already there

elip-deterministic-blinding-key.mediawiki
LeoComandini: The key is a private `bare` key, which is called `view` key.

Changed the line above to:
```diff
- Where <code>KEY</code> is a view blinding key derived as follows:
+ Where <code>KEY</code> is a ELIP-150 view blinding key derived as follows:
```

I'll add an address in the test vector later (although some test vectors do not have wildcards)

jgriffiths: Thanks. Even without wildcards, the final derived blinding key for the address will be tweaked and so IMO should be included in the test vector data.

elip-deterministic-blinding-key.mediawiki
```diff
@@ -0,0 +1,122 @@
...
+Wallets use <code>scriptPubKey</code>s starting from index 0,
+thus we choose the last non-hardened index,
+which in practice will never be used in a transaction.
+</ref>
+** Derive the <code>scriptPubKey</code> from the definite descriptor<ref>
+Here we use the <code>scriptPubKey</code> because descriptors have fields that might change without affecting the spending conditions.
+For instance changing the fingerprint of one xpub yields the same <code>scriptPubKey</code>s.
+Using the <code>scriptPubKey</code> allows us to treat such descriptors as equivalent.
+</ref>
+* Concatenate all <code>scriptPubKey</code>s obtained before.
```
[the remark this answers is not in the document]
LeoComandini: Added `OP_INVALIDOPCODE` as separator

elip-deterministic-blinding-key.mediawiki
LeoComandini: Added first address in the test vectors

elip-deterministic-blinding-key.mediawiki
LeoComandini: Done, using `elipxxxx` for now.

elip-deterministic-blinding-key.mediawiki
```diff
@@ -0,0 +1,133 @@
...
+</ref>
+* Concatenate all <code>scriptPubKey</code>s obtained before, separating them with <code>OP_INVALIDOPCODE</code>.
+* Let ''H<sub>tag</sub>'' be a [https://github.com/bitcoin/bips/blob/master/bip-0340.mediawiki#Design BIP-340 tagged hash] with tag <code>Deterministic-View-Key/1.0</code><ref>
+A tagged hash is used to make sure hashes used in this context can't be reinterpreted in another one.
+</ref>
+* The descriptor blinding key is the tagged hash of the concatenated <code>scriptPubKey</code>s, ''<code>KEY</code> = H<sub>tag</sub>(<code>scriptPubKey</code>, <code>OP_INVALIDOPCODE</code>, ...) mod n'', converted to WIF format.
+
+==Test Vectors==
+
+The following ordinary descriptor yield to the given descriptor blinding key:
```
jgriffiths, line 98: nit: `descriptor` -> `descriptors`, `key` -> `keys`

elip-deterministic-blinding-key.mediawiki
```diff
@@ -0,0 +1,133 @@
...
+
+* Test vector: 1
+** Ordinary descriptor: <code>elwpkh(xpub661MyMwAqRbcFkPHucMnrGNzDwb6teAX1RbKQmqtEF8kK3Z7LZ59qafCjB9eCRLiTVG3uxBxgKvRgbubRhqSKXnGGb1aoaqLrpMBDrVxga8/<0;1>/*)#lg37z46y</code>
+** Confidential descriptor (equivalent version): <code>ct(elipxxxx,elwpkh(xpub661MyMwAqRbcFkPHucMnrGNzDwb6teAX1RbKQmqtEF8kK3Z7LZ59qafCjB9eCRLiTVG3uxBxgKvRgbubRhqSKXnGGb1aoaqLrpMBDrVxga8/<0;1>/*))#cawe8rjm</code>
+** Confidential descriptor: <code>ct(L3TsJMSZ1gfJyLgyT16Pb8c7Tj1oxosPvZGoHpryH8frpXf6Sphb,elwpkh(xpub661MyMwAqRbcFkPHucMnrGNzDwb6teAX1RbKQmqtEF8kK3Z7LZ59qafCjB9eCRLiTVG3uxBxgKvRgbubRhqSKXnGGb1aoaqLrpMBDrVxga8/<0;1>/*))#0aphatv9</code>
+** Descriptor blinding key: <code>L3TsJMSZ1gfJyLgyT16Pb8c7Tj1oxosPvZGoHpryH8frpXf6Sphb</code>
```
jgriffiths, line 104: nit: (IMO) `Descriptor blinding key` -> `Derived descriptor blinding key`,  `Confidential descriptor` -> `Derived confidential descriptor`, then swap the order (derived  descriptor blinding key should come first).

elip-deterministic-blinding-key.mediawiki
```diff
@@ -0,0 +1,133 @@
...
+* Let ''H<sub>tag</sub>'' be a [https://github.com/bitcoin/bips/blob/master/bip-0340.mediawiki#Design BIP-340 tagged hash] with tag <code>Deterministic-View-Key/1.0</code><ref>
+A tagged hash is used to make sure hashes used in this context can't be reinterpreted in another one.
+</ref>
+* The descriptor blinding key is the tagged hash of the concatenated <code>scriptPubKey</code>s, ''<code>KEY</code> = H<sub>tag</sub>(<code>scriptPubKey</code>, <code>OP_INVALIDOPCODE</code>, ...) mod n'', converted to WIF format.
+
+==Test Vectors==
+
+The following ordinary descriptor yield to the given descriptor blinding key:
+
+* Test vector: 1
```
jgriffiths, line 100: nit: remove `:` from test vector names

elip-deterministic-blinding-key.mediawiki
LeoComandini: changed

elip-deterministic-blinding-key.mediawiki
LeoComandini: Changed

elip-deterministic-blinding-key.mediawiki
LeoComandini: done

jgriffiths:
LGTM thanks, ack.

elip-deterministic-blinding-key.mediawiki
```diff
@@ -0,0 +1,133 @@
...
+Wallets use <code>scriptPubKey</code>s starting from index 0,
+thus we choose the last non-hardened index,
+which in practice will never be used in a transaction.
+</ref>
+** Derive the <code>scriptPubKey</code> from the definite descriptor<ref>
+Here we use the <code>scriptPubKey</code> because descriptors have fields that might change without affecting the spending conditions.
+For instance changing the fingerprint of one xpub yields the same <code>scriptPubKey</code>s.
+Using the <code>scriptPubKey</code> allows us to treat such descriptors as equivalent.
+</ref>
+* Concatenate all <code>scriptPubKey</code>s obtained before, separating them with <code>OP_INVALIDOPCODE</code>.
```
apoelstra, line 90: In c52c80d9fd82f9f1471c77255c443d7096fa4e6b:

I would suggest "prefixed with `OP_INVALIDOPCODE`" rather than "separated by". That way even if there is only one scriptpubkey, an invalid opcode will still appear.

Also, is `INVALIDOPCODE` a real opcode? If you would like to pick a specific value i'd choose `OP_RESERVED` or `OP_RETURN`.

apoelstra:
c52c80d9fd82f9f1471c77255c443d7096fa4e6b lgtm. Just two minor comments.

apoelstra:
Assigned ELIP 151.

Having said this, my original vision was that people would achieve this by choosing otherwise-unused BIP32 paths and then combining their keys using `musig` or something. But given that this is currently unspecified (ELIP 150 says `Later extensions to this ELIP may specify alternate expressions...`) I think this is a simple and good alternative.

elip-deterministic-blinding-key.mediawiki
jgriffiths: `OP_INVALIDOPCODE` (0xff) is always invalid if used in scripts, from cores `src/test/script_tests.cpp`:

```
    script = ScriptFromHex("ff88ac"); // Script with OP_INVALIDOPCODE explicit
    BOOST_CHECK(!script.HasValidOps());
```

I'd avoid `OP_RETURN` but `OP_RESERVED`  would also work.

elip-deterministic-blinding-key.mediawiki
```diff
@@ -0,0 +1,133 @@
...
+<pre>
+ct(<KEY>,<DESCRIPTOR>)
+</pre>
+
+Where <code>KEY</code> is a ELIP-150 view blinding key derived as follows:
+* If the top-level script expression is <code>combo</code>, fail.
+* Future extensions that support multiple descriptors other than multi-path are not supported.
+* If <code>DESCRIPTOR</code> is multi-path, expand to every single descriptor.
+* For each single descriptor:
+** If it has wildcards, get the definite descriptor at index 2<sup>31</sup>-1<ref>
```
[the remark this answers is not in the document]
LeoComandini: Done

elip-deterministic-blinding-key.mediawiki
LeoComandini: Changed

LeoComandini:
Addressed review comments and used assigned elip number in the last commit, will squash it later

apoelstra:
Looks great!

LeoComandini:
will squash and update after #9 is merged (or abandoned/nacked)

LeoComandini:
Rebased, updated and squashed. Also updated reference impl https://github.com/ElementsProject/elements-miniscript/pull/61

apoelstra (approved):
ACK 59dae9f90e51160bec335212a63a17b21ebcc7ea

LeoComandini:
@apoelstra do we need more work/review on this? or it can be merged? related ref impl was merged https://github.com/ElementsProject/elements-miniscript/pull/61

apoelstra:
I think we can merge it.

PR #9, opened 2023-11-24
WIP: elip-0150: only allow hex bare/view blinding keys

LeoComandini:
Using key expressions as defined in BIP 380 means that descriptor
bare/view blinding keys might contain network bytes, which is not
desirable.

Note that this change also removes the possibility to have
extended keys as bare/view keys. We do so because:
* they contain network bytes too
* the extended key will never be BIP 32 derived, only the
  public/private key is tweaked
* deriving a bare/view key from a BIP 32 seed is not compatible
  with HWWs. Since they don't export private keys, we won't be
  able to get the view key and derive the output blinding keys.

If we want to store extra information on how the descriptor
blinding key was derived, we can consider extending this ELIP.

WIP until test vectors are updated.


LeoComandini:
> > * the extended key will never be BIP 32 derived, only the
> >   public/private key is tweaked
> 
> We should definitely support BIP32-deriving view keys. Supporting this was one of the primary goals of this ELIP.
> 
> If this forces us to deal with network bytes in descriptors then maybe we just have to bite the bullet.

I can rework this PR to allow for BIP32 extended keys as descriptor blinding keys.

Currently (ie ignoring this PR) a bare/view blinding key could be:
* a single private key (a WIF key)
* a single public key (66 hex chars)
* a xprv (b58 string + optional keyorigin)
* a xpub (b58 string + optional keyorigin)

Should we only change single private key to be encoded in hex rather than WIF?

Anyway this
> * deriving a bare/view key from a BIP 32 seed is not compatible
> with HWWs. Since they don't export private keys, we won't be
> able to get the view key and derive the output blinding keys.

would still apply, and that was the main rationale to remove support for them.

apoelstra:
>Should we only change single private key to be encoded in hex rather than WIF?

Yeah, I'm happy with this change. Non-BIP32 private keys are a weird edge case and I don't think it's too inconvenient not to allow WIF. Since WIF (a) has its own network byte which corresponds to a different enum than that of the xprv/xpub network; (b) has a "public key" format with no network byte at all.

>deriving a bare/view key from a BIP 32 seed is not compatible with HWWs. Since they don't export private keys, we won't be able to get the view key and derive the output blinding keys.

Ledger, for example, will export certain private keys IIRC, under a specific path for a specific vendor. Or something like that. So we shouldn't rule out the possibility that HWWs could support exporting a subset of privkeys.

LeoComandini:
> > Should we only change single private key to be encoded in hex rather than WIF?

> Yeah, I'm happy with this change.

Updated PR, text, title and related PR in elements-miniscript

jgriffiths:
> Ledger, for example, will export certain private keys IIRC, under a specific path for a specific vendor. Or something like that. So we shouldn't rule out the possibility that HWWs could support exporting a subset of privkeys.

FYI Legder allows you to fetch the blinding pubkey and the computed blinding nonces, but not any private key.

apoelstra:
Ok, 0ddbaaa2178f53e0d7287e5290368f60426576b6 looks great to me. Would like an ACK from somebody else as well.

elip-0150.mediawiki
```diff
@@ -148,8 +149,8 @@ The non-view descriptor is the same as the view descriptor except that its priva
 
 This one is a "view descriptor" using ordinary keys rather than extended ones:
 
-* View Descriptor 2: <code>ct(L3jXxwef3fpB7hcrFozcWgHeJCPSAFiZ1Ji2YJMPxceaGvy3PC1q,elwpkh(021a8fb6bd5a653b021b98a2a785725b8ddacfe3687bc043aa7f4d25d3a48d40b5))#stngzuwt</code>
-** Descriptor blinding private key: <code>L3jXxwef3fpB7hcrFozcWgHeJCPSAFiZ1Ji2YJMPxceaGvy3PC1q</code>
+* View Descriptor 2: <code>ct(c25deb86fa11e49d651d7eae27c220ef930fbd86ea023eebfa73e54875647963,elwpkh(021a8fb6bd5a653b021b98a2a785725b8ddacfe3687bc043aa7f4d25d3a48d40b5))#c2kx9zll</code>
+** Descriptor blinding private key: <code><c25deb86fa11e49d651d7eae27c220ef930fbd86ea023eebfa73e54875647963/code>
```
RCasatta, line 153: typo: position of the `<`

RCasatta:
0ddbaaa2178f53e0d7287e5290368f60426576b6 good to me (except the typo)

elip-0150.mediawiki
LeoComandini: fixed in a separate commit

apoelstra (approved):
ACK bf51db92628a59be43625af77cebe76ce8d53fb7

PR #11, opened 2024-04-15
elip150/151: test vectors: generate mainnet addresses

LeoComandini:
Consistently with the extended pubkey network.

Generated with https://github.com/ElementsProject/elements-miniscript/pull/77


apoelstra:
Spot-checked a couple to make sure that they do, in fact, come from the output of #77, but did not carefully check them all.

PR #12, opened 2024-04-23
elip150: test_vectors: add new case

LeoComandini:
With "bare" descriptor blinding key and xpub in the ordinary descriptor.


LeoComandini:
From https://github.com/ElementsProject/elements-miniscript/pull/78

Based on #11 to avoid rebase issues

apoelstra (approved):
ACK 2b7ec546c3f8e27faa8f7f93ccf8383619b2803e

PR #13, opened 2024-05-07
elip151: update spec to match reference implementation

LeoComandini:


apoelstra (approved):
ACK 06169780bafdfd700daae8b363b0231657691862

I think you could be a little more explicit by writing out `01ff` after "prefixed by its length". But you give an example so I think this is fine.

PR #16, opened 2024-06-19
init: elip for discounted CT

delta1:
first pass at ELIP for discounted CT as implemented in ElementsProject/elements#1317 and alternate implementation in ElementsProject/rust-elements#204


apoelstra:
3b286eacde2906e9fa4464bea3cffdf6d9842196 looks good to me except that it should have explicit test vectors.

delta1:
thanks @apoelstra. should they be inline or as separate files? 

what number should be used for this ELIP?

apoelstra:
I would put them inline.

This can have number 200. (For transparency -- I am just picking a number "far away" from existing BIPs which don't seem to resemble this one, and I don't have any secret table of predetermined number ranges.)

delta1:
added test vectors and used number 200

apoelstra (approved):
utACK 33e5c4d0d845941f3c595f18a4e5219ef892ee95 -- did not verify the test vectors

apoelstra:
Gonna go ahead and merge this. Can do followup PRs when people start reimplementing this, if there is feedback.

PR #21, opened 2024-09-25
elip200: fix discount weight calculation for asset and nonce

delta1:
output amount and nonce discount must be weighted by the witness scaling factor as they form part of the base transaction, not the witness data

see https://github.com/ElementsProject/elements/pull/1366 and https://github.com/ElementsProject/rust-elements/pull/218


apoelstra:
I think you should change `bytes` to `weight units` or `wu`. Multiplying a number of bytes by 4 looks wrong.

delta1:
> I think you should change `bytes` to `weight units` or `wu`. Multiplying a number of bytes by 4 looks wrong.

Thanks, done in  b80157cf02510047ad19708e8ca79d34011c3ef3

apoelstra:
Now it descrbes the commitment as beeing "33 wu" which is incorrect. It is either "33 bytes" or "132 wu". Same for the "no nonce" case.

Also the text "multiplied by 4 as part of the base transaction weight" is confusing. I'm not sure what it's supposed to mean. I guess you mean "the number of bytes is multiplied by 4 since this is non-witness data" but the number of bytes is not the subject of this sentence.

delta1:
updated in 9f08b8168328691c1bff6e0261cebbe985f39ae4

apoelstra (approved):
ACK 9f08b8168328691c1bff6e0261cebbe985f39ae4

PR #23, opened 2025-03-07
Mark ELIP 200 as active

kilrau:


apoelstra:
cc @delta1 this is correct, right?

delta1 (approved):
ACK 8d605aaef71ac99eb38b0d5b693faff34366a427

PR #24, opened 2025-03-07
elip201: Reduce Elements default dust relay feerate to match min relay feerate

kilrau:


README.mediawiki
```diff
@@ -53,6 +53,14 @@ Having an ELIP here does not make it a formally accepted standard until its stat
 | Byron Hambly
 | Standards Track
 | Draft
+| Draft
```
psgreco, line 56: Duplicated "Draft"

kilrau: Fixed

elip-0201.mediawiki
```diff
@@ -0,0 +1,36 @@
+<pre>
+  ELIP: 201
```
delta1, line 2: Numbers are normally assigned after the first draft and not chosen by the ELIP author

elip-0201.mediawiki
```diff
@@ -0,0 +1,36 @@
...
+
+This document proposes to reduce the wallet-side enforced dust limit for Confidential Transactions on the Liquid Network.
+
+===Copyright===
+
+This document is licensed under the 3-clause BSD license.
+
+===Motivation===
+
+[https://github.com/ElementsProject/elements Elements], [https://github.com/Blockstream/gdk GDK] and [https://github.com/Blockstream/lwk LWK] ("Liquid reference wallets") currently enforce a wallet-side 546 sats dust limit, originally inherited from Bitcoin Core<ref>[https://github.com/bitcoin/bitcoin/blob/ee68b05f3d639b28b9d23b48324bc5f20ccfe49c/src/policy/policy.cpp#L36 Bitcoin Core Policy]</ref>. With ELIP 200<ref>[https://github.com/ElementsProject/ELIPs/blob/main/elip-0200.mediawiki ELIP 200]</ref> live on Liquid mainnet, network fees for Confidential Transactions dropped by ~10x, with <20 sats total for a single input, single output Confidential Transaction. The default dust limit of Liquid's reference wallet libraries should be lowered to reflect this and allow for micropayment use cases on Liquid.
```
delta1, line 26: Not accurate for Elements. From the code in your policy link below, the dust limit is dependent on the output type and the node's `dustrelayfee`

README.mediawiki
```diff
@@ -53,6 +53,13 @@ Having an ELIP here does not make it a formally accepted standard until its stat
 | Byron Hambly
 | Standards Track
 | Active
+|- style="background-color: #ffcfcf"
+| [[elip-0201.mediawiki|201]]
+| Wallet
+| Dust limit reduction for Confidential Transactions
```
delta1, line 59: Reduce default dust relay fee rate in Elements

delta1:
Hi @kilrau, thanks for the submission. 

As you note, the "dust limit" for output types is node policy and unrelated to consensus of the network. 

The default `dustrelayfee` rate is 3 sats/vb (3000 sats/kvb) [1]. 

This only becomes an issue with certain Elements RPCs. As you demonstrated a wallet can trivially create confidential outputs with lower amounts. 

In my testing, a confidential WPKH output has a base size of 98 bytes. The function `GetDustThreshold` then adds a further 67.75 bytes, giving a floored total of 165 bytes. At the default dustrelayfee of 3 sat/vb this gives a dust threshold of 495 sats. 

By reducing the `dustrelayfee` on your node to 0.1 sats/vb (100 sats/kvb), your Elements RPCs can create confidential WPKH outputs with amounts as low as 17 sats ( 165 vb * 0.1 sats/vb rounded up to the next int ). 

There is an `IsDust` check in `IsStandardTx` [2], but this can only be checked for explicit transactions. 

I am not sure this requires an ELIP, personally I'd be happy to accept a PR to Elements reducing the `DUST_RELAY_TX_FEE` to match the `DEFAULT_MIN_RELAY_TX_FEE` of 0.1 sats/vb (100 sats/kvb) [3]. But if we do want to keep this information in an ELIP then I think it just needs similar motivation as above and titled something like "Reduce default dust relay fee in Elements". 

[1]: https://github.com/ElementsProject/elements/blob/master/src/policy/policy.h#L58 
[2]: https://github.com/ElementsProject/elements/blob/master/src/policy/policy.cpp#L149
[3]: https://github.com/ElementsProject/elements/blob/master/src/validation.h#L59

delta1:
@apoelstra what do you think?

kilrau:
Originally I was tasked to create this ELIP to "have something tangible to track" for the effort to lower the dust limit across the Liquid wallet ecosystem. But I am fine to reduce this to be applicable to Elements-only as proposed by @delta1 or even just close it as in that case prbly all it takes is https://github.com/ElementsProject/elements/pull/1433

delta1:
Updated as per the discussion and reference PR 

ACK 4be5ce469598f4aec55e805c2d2b1675100aa31d

psgreco:
ACK 4be5ce469598f4aec55e805c2d2b1675100aa31d

PR #27, opened 2025-09-02
Issued Asset Fees

tomt1664:
This ELIP describes proposed features for Elements that enable the use of issued assets for transaction fees in addition to the policy asset (pegged asset). 

This will enable an Elements blockchain/sidechain to use any issued asset for the transaction fees, with the assets used and rates applied being agreed and set by the block creators. Changes are proposed to relay/mempool policy and block creation. 

Several mechanisms are proposed to set, synchronise and publish the assets and rates that are accepted by the block producers.

### DIFF
--- a/README.mediawiki
+++ b/README.mediawiki
@@ -102,6 +102,13 @@ Having an ELIP here does not make it a formally accepted standard until its stat
 | Tom Trevethan
 | Standards Track
 | Draft
+|- style="background-color: #ffcfcf"
+| [[elip-0204.mediawiki|204]]
+| Applications
+| Issued Asset Fees
+| Tom Trevethan
+| Standards Track
+| Draft
 |}
 
 <!-- IMPORTANT!  See the instructions at the top of this page, do NOT JUST add BIPs here! -->
--- a/elip-0204.mediawiki
+++ b/elip-0204.mediawiki
@@ -1,0 +1,199 @@
+<pre>
+  ELIP: 204
+  Layer: Applications
+  Title: Issued Asset Fees
+  Author: Tom Trevethan <ttrevethan@blockstream.com>
+  Comments-Summary: No comments yet.
+  Comments-URI: https://github.com/ElementsProject/elips/wiki/Comments:ELIP-0204
+  Status: Draft
+  Type: Standards Track
+  Created: 2025-09-02
+  License: BSD-3-Clause
+</pre>
+
+==Introduction==
+
+===Abstract===
+
+This document proposes a modification to Elements to enable transaction fees to be paid in issued assets in addition to the <code>policyAsset</code> (or pegged asset). Individual nodes will be configured with the issued assets for fees accepted for mempool entry and relay, and at what fee rates relative to the <code>policyAsset</code>. The accepted fee assets and the relative applied fee rate can be either configured explicitly for each node, or set remotely with an authenticated on-chain transaction. Block creators can optionally publish the accepted fee assets and corresponding fee rates in the coinbase of blocks they generate to communicate this information to users and wallets. 
+
+===Copyright===
+
+This document is licensed under the 3-clause BSD license.
+
+===Motivation===
+
+Current policy rules for mempool acceptance and relay require that the <code>assetID</code> of the transaction fee is the <code>policyAsset</code>. The consensus rules however allow for any <code>assetID</code> to be used for the fee asset (so long as the transaction is otherwise valid). The current restriction limits the usefulness of the Elements platform for transacting in issued assets, as <code>policyAsset</code> must be obtained before being able to transact. Block creators may choose to accept fee payments in other assets that have value, at specified exchange rates. 
+
+Block creators can configure individual nodes explicitly with accepted assets for fees and rates, but to enable a federation to be managed with consistent fee policy (which may require frequent changes), a method to perform this configuration remotely via an on-chain transaction is also required. Individual nodes can then be configured with a 'controller' script that enables changes to the fee asset policy without requiring direct configuration of individual nodes. 
+
+In order for wallets and users to know which assets are being accepted by block creators for fees and at what rates, the block creators can publish this information in the coinbase of blocks they produce. This removes the requirement for separate servers to publish and relay this data. 
+
+The controller feature is independent of the fee rate publication feature: the controller can remotely set the fee assets and multipliers to be used as policy by the block signers. The block signers can then publish the assets and multipliers being used for users/wallets. 
+
+==Design==
+
+===Overview===
+
+Each node has complete control over which assets and at what fee rates to accept as policy. However, in practice all block creators and relaying nodes should agree to apply the same asset fee policy in order to provide a reliable service for users. The assets and fee rates accepted by any node can be set as part of the node configuration, specifying the <code>assetID</code>, the relative rate multiplier (and optionally the coinbase destination for the fee collection by the block creator) before initialization, or this can be set dynamically at runtime. 
+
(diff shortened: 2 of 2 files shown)


delta1 (commented):
First pass, a few minor things

apoelstra:
In 5f0dbbd6ca01ae824d2e779383884768b62e3cf6:

trailing whitespace on most lines

apoelstra:
In 9641693b88a3bb0c36e8a602c6ba84a8b625594a:

This introduces the terms "input" and "output" without defining them, and still says nothing about how overflow in feerates is handled.

>The minRelay rate or min mempool rate would have be bigger than ASSET_RATE_SCALE_FACTOR. But a check for this should be added.

It would just need to be greater than one. But my question was not about min relay, but about how these transactions' feerates are computed so that transactions can be ordered in the mempool (AFAICT this is never mentioned in this document).

tomt1664:
Sorry for the delay in replying. I've gone through and changed the logic of how the rate multiplier is applied, changing it to computing an 'effective rate' for a transaction with an issued asset fee which can then be used in place of the policyAsset rate for both meeting relay/mempool minimums and mempool ordering. 

> This introduces the terms "input" and "output" without defining them, and still says nothing about how overflow in feerates is handled.

I've removed that, and specified clearly the format of values used in configuration and RPC arguments and how it is then converted to the fixed precision integer. 

>and still says nothing about how overflow in feerates is handled. 

> It would just need to be greater than one. 

Yes, you are right. I've added a maximum value parameter for the rate multiplier. 
In the calculation for the new effective fee rate, the rate multiplier is multiplied by the fee rate in the issued asset. This could potentially overflow int64, but only if the fee rate in the issued asset multiplied by the multiplier gives an effective fee rate of > 922 * COIN (a value of 922 LBTC/vbyte). It's implausible a rate multiplier and fee would be set to this level on purpose, but to protect against that possibility, if the result of the multiplication does exceed that value, then the effective fee rate is set to this maximum. 

> But my question was not about min relay, but about how these transactions' feerates are computed so that transactions can be ordered in the mempool (AFAICT this is never mentioned in this document). 

The 'effective rate' for an issued asset fee transaction is now used for both comparison with minimums and also for mempool ordering. Statements explaining this have been added to the text.

elip-0202.mediawiki
```diff
@@ -0,0 +1,232 @@
...
+
+For fees outputs in an issued asset, the fee rate (in units/vbyte) will be scaled by a fee rate multiplier to determine an 'effective rate'. This effective rate is then used in place of the policyAsset fee rate for the purpose of meeting relay and mempool minimums and for mempool ordering. 
+
+E.g. An issued asset is USDT. The current conversion rate between USDT and LBTC (<code>policyAsset</code>) is 100:1 (i.e. 100 USDT to one LBTC) the issued asset fee rate multiplier would then be set to 0.01 (i.e. the fee paid in USDT will be multiplied by 0.01 for comparison with LBTC fees). If the USDT asset is used for the fee in any transaction, it should pay 100 times in USDT base units the fee rate that would have applied in LBTC (the policyAsset). If the current minimum fee rate for mempool and relay is 1 sat/vbyte in LBTC (i.e. the policy asset fee rate), then the effective minimum fee rate for this asset would be 100 units/byte. The minimum fee output value required for a transaction of size e.g. <code>vsize = 257</code> transaction to be relayed would be: <code>257 x 1 x 100 = 25700</code> base units of the USDT asset.
```
apoelstra, line 44: In 4b5eeb206931975787cb5d67114ecd4ed8d1557c:

I don't undersztand this example. For one thing, it seems cherry-picked to have an integer ratio; what if the actual exchange rate is `105:100` say? For another thing, it seems to normalize the fee to USDT rather than L-BTC. If there are multiple fee outputs with different assets then which one is chosen to normalise to?

elip-0202.mediawiki
```diff
@@ -0,0 +1,232 @@
...
+
+In order for the acceptance of issued asset and rates for fees to be reliable and predictable, all block creators and bridging nodes (and all relaying nodes) should be applying the same policy with the same list of accepted assets and rates. To coordinate this and apply changes to a large number of nodes is impractical, which may be required regularly as exchange rates change. To provide an alternative method for remotely modifying the accepted assets and rates, individual nodes can be configured to accept updates via an on-chain transaction - this way all block creator and bridge nodes can have their applied policy updated simultaneously without requiring direct access to the RPC interface.
```
apoelstra, line 48: In 4b5eeb206931975787cb5d67114ecd4ed8d1557c:

This paragraph doesn't make sense after the previous paragraph. It says that the blocksigners will encode exchange rates in `OP_RETURN` outputs from coinbases, then describes an "alternate" method in which exchange rates are accepted "via an on-chain transaction" with no further details given. Is this the same thing as the OP_RETURN mechanism you just described? It is contrasted to "requiring direct access to the RPC interface" but you've only described using the RPC interface to access the current rates, not for setting them.

elip-0202.mediawiki
tomt1664: A simple integer ratio was chosen to make the arithmetic of this introductory example easier, but maybe it's too abstract. A more realistic exchange rate can be used, which illustrates precision/rounding issues. 

The fee is paid in USDT, and the monetary value of that fee must equal (or exceed) the monetary value of the fee currently required in LBTC. The multiplier converts the monetary value of an amount of USDT to an amount of LBTC. 

Another attempt:

```
E.g. An issued asset is USDT. The current accepted exchange rate between USDT and LBTC (the <code>policyAsset</code>) is $114,171.23 to 1 LBTC. The issued asset fee rate multiplier for this ratio is then set to <code>1/114,171.23 = 0.00000876</code> (rounded to a precision of 1e-8). For a transaction that uses the USDT asset for the fee, the fee rate (in units/vbyte) multiplied by this number must be greater than or equal to the minimum fee rate (in sat/vbyte) applied in LBTC. 

If the current minimum relay fee rate is 1 sat/vbyte in LBTC, then the minimum fee rate required in USDT would be <code>1/0.00000876 = 114156</code> units/vbyte (rounded up to the nearest integer). To relay this transaction, a client will determine the fee rate in the USDT asset, multiply it by 0.00000876 to calculate the effective rate and then verify it is greater than or equal to 1 sat/vbyte. 
```

In the case of multiple fee outputs, in principle, an effective fee (the fee amount scaled by the multiplier for each different asset) can be calculated for each fee output and then an effective fee rate for the transaction determined. However, it may be simpler to require wallets to use a single asset and output, and only consider the first one.

elip-0202.mediawiki
tomt1664: > This paragraph doesn't make sense after the previous paragraph. It says that the blocksigners will encode exchange rates in `OP_RETURN` outputs from coinbases, then describes an "alternate" method in which exchange rates are accepted "via an on-chain transaction" with no further details given. Is this the same thing as the OP_RETURN mechanism you just described?

These are two separate features. The first paragraph introduces a method for publishing the assets and rates accepted by a node. The second is a separate method for remotely *setting* the assets and rates accepted by a node without direct access to it, as an alternative to configuring node directly. 

> It is contrasted to "requiring direct access to the RPC interface" but you've only described using the RPC interface to access the current rates, not for setting them.

It states above "This can be set as part of the node configuration, specifying the <code>assetID</code>, the relative rate multiplier (and optionally the coinbase destination for the fee collection by the block creator) before initialization, or set dynamically at runtime via a new RPC option."

delta1 (approved):
preliminary utACK bcd0cb8620e8e98dfded56342fc5e523bdbd1a63; some minor comments/suggestions

psgreco:
Carrying over @delta1 's ack after the nits/rebase

PR #28, opened 2025-09-25
ELIP 202: Optional sidechain peg-in subsidy and minimum peg-in amount

delta1:
Initial draft of new ELIP for option sidechain peg-in subsidy and minimum peg-in amount. 

Still todo: link to reference implementation and add test vectors.


elip-0xxx.mediawiki
```diff
@@ -0,0 +1,150 @@
...
+
+New mempool validation is implemented to check for a peg-in subsidy from the subsidy height. This requires that a peg-in transaction, whose value is below the subsidy threshold, must contain one or more <code>OP_RETURN</code> outputs burning the value of spending a P2WSH output on the parent chain at the same fee rate as the parent peg-in. A new policy will also reject peg-ins below the minimum peg-in amount after the minimum peg-in height.
+
+This validation requires that the sidechain node is validating peg-ins against the parent node, as the sidechain node calls <code>getrawtransaction</code> to retrieve the fee paid and vsize of the parent transaction to calculate its fee rate. If the sidechain node is not validating peg-ins (<code>validatepegin=0</code>), then it will accept the transaction without subsidy check, but this transaction will fail to be relayed to nodes that are validating peg-ins. The minimum acceptable parent fee rate is 1 sat/vb.
```
apoelstra, line 49: In 10a67c39a41e08dd11e6ca346a28313db108a8e1:

Why have this rule? If a pegin was accepted on the Bitcoin chain at a lower feerate then why are we second-guessing that? Also, if we want to enforce a minimum, why set it by forbidding certain parent transactions (which I guess would then just burn coins, or...?) instead of setting an alternate minimum on the size of the OP_RETURN output?

PR #33, opened 2026-03-12
Add ELIP: Deterministic Derivation of Application Keys

miketlk:
Initial draft of a new ELIP defining a deterministic scheme for deriving application-specific keys on hardware wallets.

The proposal reserves a dedicated BIP-32 subtree `m / 4280400' / 4932953' / 1'` (corresponding to "APP"/"KEY"/v1) for the application root [private] key. Applications can then derive labeled keys from this root using HMAC-SHA256.

The goal is to provide a simple and deterministic mechanism for deriving multiple application keys, including a wallet policy registration key, without reusing existing wallet/account derivation paths. It's also an alternative to SLIP-0021 (hierarchical derivation of symmetric keys) which has limited support among existing hardware wallets.

Still TODO: add reference implementation and test vectors.

### DIFF
--- a/elip-app-key.mediawiki
+++ b/elip-app-key.mediawiki
@@ -1,0 +1,170 @@
+<pre>
+  ELIP: ?
+  Layer: Applications
+  Title: Deterministic Derivation of Application Keys
+  Author: Mikhail Tolkachev <contact@miketolkachev.dev>
+  Comments-Summary: No comments yet.
+  Status: Draft
+  Type: Standards Track
+  Created: 2026-03-10
+  License: BSD-2-Clause
+</pre>
+
+==Abstract==
+
+This document defines a deterministic method for deriving application-specific keys from a dedicated BIP-32 derivation path.
+
+The scheme enables wallet applications to derive internal keys for multiple purposes from a common root key, using a construction that is independent from other key hierarchies.
+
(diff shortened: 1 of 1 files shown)


jgriffiths:
BIP85 seems like it covers this use case? It's implemented by Jade and Coldcard at least already.

miketlk:
@jgriffiths , thanks for the feedback. That's a very good point. I really overlooked BIP85 because I remembered it as a way to derive child mnemonics from a seed, as it's commonly presented.

After checking the spec, it actually fits this use case quite well, including deterministic derivation of symmetric application keys. So this ELIP may indeed be reinventing functionality that BIP85 already provides. 

At this point, it probably makes more sense to standardize a specific BIP85-based derivation for the wallet policy registration key (application number + entropy interpretation) instead of introducing a completely new derivation scheme.

apoelstra:
Should I close this PR and we'll revisit in the form of a "how to use BIP 85" ELIP?

miketlk:
I think this PR should be closed for sure as unnecessary.
But I'm not certain about do we need a new ELIP describing how to use BIP85 to derive wallet policy registration key, or it's better simply add to documentation of Liquid app for Ledger. It depends on whether it has reuse potential for other hardware wallets or apps.

PR #35, opened 2026-05-13
Add ELIP: Wallet ABI Transaction Creation Protocol

KyrylR:
## Motivation (copied from the text) 

Applications need a standard way to ask a wallet to construct and sign a transaction from an application-level intent
while keeping wallet-owned UTXOs private to the wallet. This protocol defines a communication schema that gives the requester a signing/construction capability without requiring the wallet to share its UTXO set, balances, or wallet-internal selection state.


delta1:
Assigned ELIP 208

PR #36, opened 2026-06-01
Add ELIP: Liquid Wallet RPC Profile

KyrylR:
This ELIP defines a Liquid wallet JSON-RPC profile for the wallet SDKs and other dapp-to-wallet transports.
It specifies how a dapp identifies a connected Liquid account, requests public wallet descriptors, reads wallet-computed
balances and UTXOs, requests transfers, requests PSET signatures, signs messages, uses deterministic identity keys,
delegates complex confidential transaction construction to the Wallet ABI transaction protocol, and receives descriptor-change events.

---

Bitcoin integrations can use public addresses and PSBTs as their main wallet-facing primitives.
Liquid wallets require a more explicit profile because Liquid has confidential transactions, multiple issued assets,
network-specific policy assets, CT descriptors, blinding keys, PSETs, and Simplicity.

Dapps need enough public wallet information to identify the connected account, construct descriptor-aware receive
and change logic, read user-approved balances and UTXOs, and build valid PSET signing requests. At the same time,
wallets must not leak view material merely to support these flows. This ELIP separates public descriptor disclosure
from wallet-computed balance and UTXO disclosure, and it defines a descriptor-format negotiation mechanism
based on descriptor representations.

### DIFF
--- a/elip-liquid-wallet-rpc-profile.mediawiki
+++ b/elip-liquid-wallet-rpc-profile.mediawiki
@@ -59,13 +59,15 @@ The following terms are used throughout this ELIP.
 |-
 | Public wallet descriptor || An ordinary Liquid/Elements output descriptor with public spend-key material and no <code>ct(...)</code> wrapper. It can derive scriptPubKeys and unconfidential addresses. It cannot derive confidential addresses, unblind outputs, or spend funds.
 |-
-| Account identifier || The profile-specific account suffix selected by this ELIP: the first external unconfidential address at index <code>0</code> derived from the public wallet descriptor.
+| Account identifier || The [https://github.com/ElementsProject/ELIPs/blob/main/elip-0144.mediawiki ELIP-0144] Liquid account ID for the connected wallet account: <code>chain_id + ":" + dwid</code>. This value is used in <code>accountIdentifier</code> result fields and in <code>account</code> request parameters.
 |-
-| Connected account || The account string displayed and persisted by the dapp: <code>chain_id + ":" + account_identifier</code>.
+| DWID || The Deterministic Wallet IDentifier defined by [https://github.com/ElementsProject/ELIPs/blob/main/elip-0152.mediawiki ELIP-0152]. In this profile, the DWID component of an account identifier MUST identify the same wallet descriptor set as the public wallet descriptor returned for the connected account.
+|-
+| Connected account || The account identifier authorized for the active transport session.
 |-
 | Public confidential descriptor || An [https://github.com/ElementsProject/ELIPs/blob/main/elip-0150.mediawiki ELIP-0150] CT descriptor with public blinding-key material only. It can derive confidential addresses. It cannot derive private blinding keys and cannot unblind outputs.
 |-
-| View descriptor || A CT descriptor that lets the holder derive private blinding keys or otherwise unblind wallet outputs. Examples include descriptors using [https://github.com/satoshilabs/slips/blob/master/slip-0077.md SLIP-0077] private master blinding material, private descriptor blinding keys, descriptor blinding xprvs, or [https://github.com/ElementsProject/ELIPs/blob/main/elip-0151.mediawiki ELIP-0151] <code>ct(elip151, &lt;DESCRIPTOR&gt;)</code>. View descriptors MUST NOT be returned by this profile.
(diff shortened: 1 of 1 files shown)


delta1:
Assigned ELIP 207

elip-liquid-wallet-rpc-profile.mediawiki
```diff
@@ -0,0 +1,1144 @@
...
+signing request is already authorized by the request context.
+
+{| class="wikitable"
+! Parameter !! Type !! Required !! Description
+|-
+| <code>identity</code> || String || Yes || The identity URI shown to the user.
+|-
+| <code>curve</code> || String || Yes || Supported value: <code>nist256p1</code>.
+|-
+| <code>challenge</code> || String || Yes || The challenge bytes as lowercase hex without <code>0x</code>. The byte length MAY depend on the wallet implementation.
```
stringhandler, line 740: I suggest that a fixed prefix is always added to this challenge, e.g. "signIdentity", so that the identity is not used to sign some transactions, e.g. a p2pk

stringhandler: Update, I see that there is signMessage later that does this.

I would flag this method at highly dangerous and should potentially be removed from this ELIP and have it's own ELIP

elip-liquid-wallet-rpc-profile.mediawiki
```diff
@@ -0,0 +1,1144 @@
...
+    "curve": "nist256p1",
+    "index": 0,
+    "type": "slip-0013",
+    "publicKey": "04012c4f72037ddf467581dd8081bd365b668c9e00fdec2a9c9ff7f4a23df3ff747cd1fb99256b631c6316d1f1b692df829d3c6a687d327d3a3942822499a2abc1",
+    "signature": "00c3fec3b39151ec274f7fb821e6b019c02625e2bfea74fca6635b19d28bef9de53fd7937ca7b10e6be7e228ae96923a45dd4add2aa94920fe660b522b6a9bf6c2"
+  }
+}
+</pre>
+
+===<code>signPset</code>===
```
stringhandler, line 794: This method also seems incredibly dangerous as the wallet may not always be able to determine exactly what the PSET is doing. 
As with my comment on `signIdentity` above, I would suggest removing it from this ELIP and creating a new ELIP just for these methods

elip-liquid-wallet-rpc-profile.mediawiki
```diff
@@ -0,0 +1,1144 @@
...
+    "identity": "ssh://jade@blockstream.com",
+    "curve": "nist256p1",
+    "index": 0,
+    "type": "slip-0013",
+    "publicKey": "04012c4f72037ddf467581dd8081bd365b668c9e00fdec2a9c9ff7f4a23df3ff747cd1fb99256b631c6316d1f1b692df829d3c6a687d327d3a3942822499a2abc1"
+  }
+}
+</pre>
+
+===<code>getIdentitySharedKey</code>===
```
stringhandler, line 642: While the KDF info does help, I am worried about the wallet showing a random string to the user. This method also provides a secret key directly to the website, which I think is dangerous. In general, even shared secrets should not leave the wallet

stringhandler (commented):
Sounds great, apart from methods I've highlighted. I would suggest moving these controversial methods to a separate ELIP

elip-liquid-wallet-rpc-profile.mediawiki
```diff
@@ -0,0 +1,1144 @@
...
+|-
+| <code>chain_id</code> || The [https://github.com/ElementsProject/ELIPs/blob/main/elip-0144.mediawiki ELIP-0144] chain identifier: <code>"bip122:" + first_32_lowercase_hex_characters_of_genesis_block_hash</code>.
+|-
+| Public wallet descriptor || An ordinary Liquid/Elements output descriptor with public spend-key and no <code>ct(...)</code> wrapper. It can derive scriptPubKeys and unconfidential addresses. It cannot derive confidential addresses, unblind outputs, or spend funds.
+|-
+| Account identifier || The [https://github.com/ElementsProject/ELIPs/blob/main/elip-0144.mediawiki ELIP-0144] Liquid account ID for the connected wallet account: <code>chain_id + ":" + dwid</code>. This value is used in <code>accountIdentifier</code> result fields and in <code>account</code> request parameters.
+|-
+| DWID || The Deterministic Wallet IDentifier defined by [https://github.com/ElementsProject/ELIPs/blob/main/elip-0152.mediawiki ELIP-0152]. In this profile, the DWID component of an account identifier MUST identify the same wallet descriptor set as the public wallet descriptor returned for the connected account.
+|-
```
lilbonekit, line 67: Dapps still need a real address to show before a deposit. Right now the only way is deriving it ourselves from the descriptor. LWK's own `wollet.address(None)` already gives the last-unused external address – could that be exposed as a read-only `getReceiveAddress`, display-only, no spend capability?

PR #40, opened 2026-06-30
Add ELIP: Clear Signing for Liquid Wallets

KyrylR:
The core intention is to provide an exhaustive way/algorithm for wallets to explain to the user what they are signing with clear trust assumptions


delta1:
Assigned ELIP 206

PR #41, opened 2026-07-03
Add TX manifest proposal

stringhandler:
Adds an ELIP for a new TX manifest format for wallets. Specifically aimed at making SimplicityHL contracts easier to integrate into wallets.

### DIFF
--- a/README.mediawiki
+++ b/README.mediawiki
@@ -109,6 +109,13 @@ Having an ELIP here does not make it a formally accepted standard until its stat
 | Tom Trevethan
 | Standards Track
 | Draft
+|- style="background-color: #ffcfcf"
+| [[elip-0205.mediawiki|205]]
+| Wallet
+| Transaction Manifests for Simplicity Contracts
+| stringhandler
+| Standards Track
+| Draft
 |}
 
 <!-- IMPORTANT!  See the instructions at the top of this page, do NOT JUST add BIPs here! -->
--- a/elip-0205.mediawiki
+++ b/elip-0205.mediawiki
@@ -1,0 +1,1018 @@
+<pre>
+  ELIP: 205
+  Layer: Wallet
+  Title: Transaction Manifests for Simplicity Contracts
+  Author: stringhandler <stringhandler@protonmail.com>
+  Comments-Summary: No comments yet.
+  Comments-URI: TBD
+  Status: Draft
+  Type: Standards Track
+  Created: 2026-07-02
+  License: BSD-3-Clause
+</pre>
+
+==Introduction==
+
+===Abstract===
+
+This ELIP defines the '''transaction manifest''', a JSON document that describes the transactions of a multi-step protocol on Elements: which UTXOs each transaction consumes, which outputs it creates, and which witnesses are needed to satisfy each input. A wallet that understands the format can execute any protocol described by a manifest, without hard-coded knowledge of that protocol.
+
+The format was designed for driving [https://github.com/BlockstreamResearch/simplicity Simplicity] contracts, where a spend must often recreate covenant outputs with particular amounts, assets and scripts, and where existing descriptor and PSET tooling gives no help with transaction construction. It is not limited to Simplicity, or to covenants, but they are the main use case.
+
+A manifest declares the covenant scripts (''UTXO types'') a contract uses, optional ''contract templates'' that group typed fields with the ''actions'' callable against them, and the actions themselves. Two companion files carry per-deployment data: an ''instance file'' holding the field values fixed when a contract was created, and a ''state file'' tracking the contract's live on-chain UTXO set.
+
+Alongside the transaction description, a manifest carries the signer-facing text a wallet shows before broadcast — an action's <code>intent</code> line and per-leg <code>ui</code> labels — and defines a canonical form so a manifest can be hashed, published and signed under a stable identifier.
+
+===Copyright===
+
+This document is licensed under the 3-clause BSD license.
+
+===Motivation===
+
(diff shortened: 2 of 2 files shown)


elip-xxx.mediawiki
```diff
@@ -0,0 +1,552 @@
...
+}
+</source>
+
+{| class="wikitable"
+! <code>type</code> !! Resolves to !! Derivation path
+|-
+| <code>"wallet_key"</code> || 32-byte x-only BIP340 Schnorr pubkey (hex) || <code>m/86h/&lt;coin&gt;h/0h/0/0</code>
+|}
+
+<code>coin</code> = 0 for mainnet, 1 for testnet.
```
Arvolear: I believe it's `0` for Bitcoin and `1776` for Liquid.

elip-xxx.mediawiki
```diff
@@ -0,0 +1,552 @@
...
+|-
+| <code>amount_sat</code> || number or string || YES || Output amount in satoshis; formula or literal
+|-
+| <code>asset</code> || string || YES || Asset ID; same syntax as the input <code>asset</code> field
+|-
+| <code>required_index</code> || integer || NO || Required transaction output index (same semantics as input <code>required_index</code>)
+|-
+| <code>optional</code> || boolean || NO || If <code>true</code>, output may be omitted. Default: <code>false</code>.
+|-
+| <code>confidential</code> || boolean || NO || Whether this output is blinded (Liquid/Elements only). Overrides the chain default (see Blinding below).
```
Arvolear: Can this possibly be derived from the `destination` field? The provided address format will say if confidentiality is required.

elip-xxx.mediawiki
```diff
@@ -0,0 +1,552 @@
...
+
+'''<code>"simplicityhl"</code> (static SimplicityHL value).''' Fields: <code>value</code> (required; a SimplicityHL value expression: <code>"0x&lt;hex&gt;"</code> for byte arrays, <code>"Left(())"</code> / <code>"Right(())"</code> for Either types) and <code>simplicity_type</code>
+
+'''<code>"Signature"</code> (BIP340 Schnorr signature).''' Fields: <code>sig_type</code> (required; currently only <code>"simplicity.sig_all_hash"</code>) and <code>source</code> (required; <code>{"type": "wallet", "key": &lt;reference&gt;}</code> where the reference resolves to a 64-hex-char x-only public key). The tool computes the sighash, signs with the wallet key matching the given pubkey, and injects the 64-byte signature as a <code>"simplicityhl"</code> witness. The value <code>"simplicity.sig_all_hash"</code> is namespaced to make clear it is Simplicity's own <code>sig_all_hash</code> jet commitment over the full transaction (computed via the Elements transaction environment), '''not''' the standard Elements SIGHASH_ALL.
```
Arvolear: Sometimes signatures may not have standalone witness names and instead have to be injected into tuples and tuples of tuples.

elip-xxx.mediawiki
```diff
@@ -0,0 +1,552 @@
...
+  "chain": "liquid",
+  "utxo_types": {
+    "p2pk_output": {
+      "script": { "type": "simplicity", "source": "./p2pk.simf" },
+      "asset": "lbtc"
+    }
+  },
+  "actions": {
+    "Pay": {
+      "params": {
```
Arvolear: I don't fully understand. Do "params" get populated by the wallet (prompting the user for some input), or instead taken from the "instance/build file" passed alongside the manifest?

stringhandler: Yes, they are shown to the user. Would it make more sense to name these "args"?

elip-xxx.mediawiki
stringhandler: This is true. In my example I changed the contract to have two separate witness fields, but let me try get a working example as you say

elip-xxx.mediawiki
stringhandler: The address might be a simplicity covenant, in which case the address is the same for both. 

Also, the destination might be "wallet" which allows the wallet to create an address, although I don't see a need for turning off confidential amounts for self addresses.

elip-xxx.mediawiki
Arvolear: Here is the injector code we use in simplex if you need some inspiration:

https://github.com/BlockstreamResearch/smplx/blob/master/crates/sdk/src/signer/wtns_injector.rs

elip-xxx.mediawiki
Arvolear: Do we expect the wallet to handle this prompting? It may be quite arbitrary and complex. I thought the dapp itself (the website) would be used to navigate the user, and then send all the data together with the manifest to the wallet just for the "confirmation" of the action.

elip-xxx.mediawiki
```diff
@@ -0,0 +1,552 @@
...
+The key words "MUST", "MUST NOT", "SHOULD", and "MAY" are to be interpreted as described in RFC 2119.
+
+====Top-Level Structure====
+
+<source lang="json">
+{
+  "manifest_version": "0.1.0",
+  "protocol": "p2pk-simplicity",
+  "description": "...",
+  "chain": "elements",
```
Arvolear: Also, I'd identify the chain according to BIP-122 and [ELIP-144](https://github.com/ElementsProject/ELIPs/blob/main/elip-0144.mediawiki).

stringhandler: agreed

delta1:
Let's assign ELIP 205

PR #45, opened 2026-08-04
elip150: add multi-key test vectors and forbid paths in blinding key

### LINKED ISSUE #34: Confidential Descriptors lack multi-key test cases
Two related issues in https://github.com/ElementsProject/ELIPs/blob/main/elip-0150.mediawiki:

Firstly , all examples/test cases use single (non-blinding) key expressions - i.e. no there are index-derivable keys such as `KEY/*`, `KEY/<0;1>/*` etc.

These are likely the most common use of such descriptors, with single key use-cases limited to sharing single addresses for counterparty address unblinding while multi-key expressions are what wallets will register and use in practice. It would be helpful to have a few such test cases derived at some index `n` for verifying implementations.

Secondly, it seems obvious from the way non-slip77 blinding keys are derived that the blinding key itself (where given as a `KEY` expression) must not contain a path expression. However this isn't explicit and there are no negative test cases where a blinding key contains a path.

delta1:
Add Valid Descriptors 11-12 covering multisig DESCRIPTOR expressions with index-derivable keys (KEY/0/* and BIP389 multipath KEY/<0;1>/*), and Invalid Descriptor 8 showing a blinding key with a derivation path is rejected. Also make explicit in the spec text that BLINDING_KEY must resolve to a single fixed key with no path component. Closes #34.


elip-0150.mediawiki
```diff
@@ -51,6 +51,8 @@ Here <code>DESCRIPTOR</code> refers to any existing descriptor, e.g. <code>elwsh
 
 In the <code><KEY></code> expression, all key(s) must be compressed; x-only and uncompressed keys are invalid.
 
+The <code><KEY></code> expression must not contain a derivation path, whether hardened, unhardened, or a wildcard (e.g. <code>/0</code>, <code>/0h</code>, or <code>/*</code>). The same blinding key is used, independently tweaked per output as described below, for every address generated by <code><DESCRIPTOR></code>, including all addresses produced by a wildcard <code><DESCRIPTOR></code>; there is no notion of deriving a different blinding key per address. <code><KEY></code> must therefore resolve to a single, fixed public or private key.
```
jgriffiths: > there is no notion of deriving a different blinding key per address

There really _is_ such a notion (the key is tweaked per address); I think you want to word this differently.

Agree that hardened and wildcard elements should be disallowed. Also, multi-path `<0;1>` should be disallowed along with any future path elements that change the solved cardinality of the key (i.e. that resolve to >1 key).

> must therefore resolve to a single, fixed public or private key.

I don't see an issue with the key itself having a fixed derivation path of unhardened elements, and can see some use cases where that might be useful. What is the rationale for disallowing it, given that the final blinding key can always be derived for both view and bare keys, and the above constraint is satisfied in this case?

elip-0150.mediawiki
```diff
@@ -138,6 +140,36 @@ The following CT descriptors should be parseable and generate the given addresse
...
+** SLIP77 master blinding key: <code>df7fe47acf54f647a1ecaa1523234994d7119185c25baf2ace7b0d732c7d5305</code>
+** At index 0:
+*** Confidential address: <code>lq1qqdcfjuf8q2uwgjrglewrpw8g4h72crxhuj4c9qgykyzhvzee8juhq252juuy9v8cfc5w640pvfdh03tktnx5kqacduxl4lamk9f66epwj78mrmumgak7</code>
+*** Unconfidential address: <code>ex1q929fwwzzkruyu28d2hskykmhc4m9en2tqwux7r06l7amz5advshqajz2dg</code>
+** At index 1:
+*** Confidential address: <code>lq1qq0sl3wlghq0ekhgswvz329gudf05splg8grv94u4qhkug8cjt5w467f24cugvcm3racqvs4ymcfqchh58ryqu0jwvchgwynhlh0dwsul0aavnf9y0ltd</code>
+*** Unconfidential address: <code>ex1q0y42uwyxvdc37uqxg2jduysvtm6r3jqw8e8xvt58zfmlmhkhgw0sxz2qmk</code>
+** At index 5:
+*** Confidential address: <code>lq1qq2qv4ekf265fcr3wzmlxg3zaxsp79c9r373egh9x7h56v8chlm4eqeqmd6l7z264sk8zu556w4wvyp99ps5yfp0f549e95xdv7ea2xv645tv7ru9j9km</code>
+*** Unconfidential address: <code>ex1qvsdka0lp9d2ctr3w22d82hxzqjjsc2zysh562juj6rxk0v74rxdq2qmc5n</code>
```
jgriffiths, line 162: Please provide the generated scripts in addition to the address; this helps when implementing.

elip-0150.mediawiki
```diff
@@ -186,6 +218,9 @@ Finally, the following are invalid test vectors that should not be parseable:
 * Invalid Descriptor 7
 ** <code>ct(L3jXxwef3fpB7hcrFozcWgHeJCPSAFiZ1Ji2YJMPxceaGvy3PC1q,elwpkh(03774eec7a3d550d18e9f89414152025b3b0ad6a342b19481f702d843cff06dfc4))#gcy6hcfz</code>
 ** Reason: single private key is WIF, not hex
+* Invalid Descriptor 8
+** <code>ct(xpub6ERApfZwUNrhLCkDtcHTcxd75RbzS1ed54G1LkBUHQVHQKqhMkhgbmJbZRkrgZw4koxb5JaHWkY4ALHY2grBGRjaDMzQLcgJvLJuZZvRcEL/0/*,elwpkh(xpub69H7F5d8KSRgmmdJg2KhpAK8SR3DjMwAdkxj3ZuxV27CprR9LgpeyGmXUbC6wb7ERfvrnKZjXoUmmDznezpbZb7ap6r1D3tgFxHmwMkQTPH))#t8tpvh0m</code>
+** Reason: blinding key expression must not contain a derivation path
```
jgriffiths: please add each listed forbidden element as its own test case (and update with a valid test case assuming fixed unhardened paths are allowed).

elip-0150.mediawiki
delta1: Thanks for the feedback, updated. 

Reworded so the constraint is about how the blinding key is obtained, with the per-output tweak called out explicitly.

Adopted your framing, the rule is now stated as the general principle rather than enumerating specific forms, so it covers multipath and anything similar added later.

Agreed on fixed unhardened path. Added Valid Descriptor 13 covering exactly that case: a fixed `/0/1` blinding key wrapping a wildcard descriptor, which doubles as a demonstration of the single-key-tweaked-per-output behaviour.

Hardened elements stay disallowed, with the rationale now in the text: they can't be derived from an extended public key, so an xprv blinding key would stop generating the same addresses as the corresponding xpub — breaking the view/non-view equivalence the existing test vectors demonstrate.

delta1: Found an issue: Disallowing wildcards conflicts with current rust-elements-miniscript behaviour, which is intentional and has a regression test. `confidential::Descriptor::at_derivation_index` applies the index to the blinding key as well as to the descriptor, and the `descriptor_wildcard` test asserts that `ct(xpub/*,elwpkh(xpub/*))` at index 1 yields a specific address. https://github.com/ElementsProject/elements-miniscript/blob/80dc9859f0329c51968b0ce53a7cd328942a7769/src/confidential/mod.rs#L577

There is a `"wildcard blinding key"` error in `Key::to_public_key`, but going by the comment in that same test it's only a safety net for the case where a `Descriptor<DefiniteDescriptorKey>` still textually holds a wildcard blinding key (`Key` isn't generic over `Pk`, so the invariant can't be expressed in the type), it isn't a policy rejection.

So for that implementation this is a breaking change: the test would need to be inverted or removed, and any wallet relying on per-index blinding keys would be affected. The alternative is to define wildcard semantics (blinding key derived at the same index) instead of forbidding them, but that goes against the "exactly one key" principle.

Separately, hardened blinding keys currently panic rather than error in elements-miniscript — `ToPublicKey::to_public_key` unwraps the `HardenedChild` error returned by `derive_public_key`, reached via `bare::tweak_key`. That looks like a bug independent of this ELIP; which I'll open a PR for.

jgriffiths: Thanks, that's all very illuminating. I am ok with allowing a single wildcard if it is documented to derive at the same index as the corresponding descriptor portion.

jgriffiths: Might be worth documenting that a derived blinding key (with or without a wildcard) does not lead to blinding key re-use because the key tweak depends on the generated script.

elip-0150.mediawiki
```diff
@@ -220,7 +243,13 @@ Finally, the following are invalid test vectors that should not be parseable:
 ** Reason: single private key is WIF, not hex
 * Invalid Descriptor 8
 ** <code>ct(xpub6ERApfZwUNrhLCkDtcHTcxd75RbzS1ed54G1LkBUHQVHQKqhMkhgbmJbZRkrgZw4koxb5JaHWkY4ALHY2grBGRjaDMzQLcgJvLJuZZvRcEL/0/*,elwpkh(xpub69H7F5d8KSRgmmdJg2KhpAK8SR3DjMwAdkxj3ZuxV27CprR9LgpeyGmXUbC6wb7ERfvrnKZjXoUmmDznezpbZb7ap6r1D3tgFxHmwMkQTPH))#t8tpvh0m</code>
-** Reason: blinding key expression must not contain a derivation path
+** Reason: blinding key expression must not contain a wildcard, as it does not resolve to exactly one key
```
jgriffiths, line 246: Nit: this is a hardened element `0h`, and the case below is a multi-index `<0;1>` - you've documented as wildcard then hardened element.

